PR #91, opened 2015-08-31
Fix an off-by-1 error

lwf:
To make parsing of consistent hash hosts correct.


rtkrruvinskiy:
This was definitely an oversight on my part when I switched to using `Count`.

PR #116, opened 2016-05-13
Add prepend and append options for destinations

tcolgate:
This is an preliminary attempt to add support for prepending, and
appending, some arbitrary strings to the metrics name of before
relaying onward.

  I'm raising this to see if you are at all interested in this, or
if you think this should part of the relay at all. There will be
a minor performance impact when the settings are not used.

  Prepending should be reasonably efficient.

  Appending will create a fair bit of garbage in the current
implementation.

  I've not benchmarked either option yet, but can do if you are
interested in integrating something like this.

  I've also not added the web interface support yet.

### DIFF
--- a/carbon-relay-ng.ini
+++ b/carbon-relay-ng.ini
@@ -24,12 +24,13 @@ legacy_metric_validation = "strict"
 # here's some examples:
 init = [
      'addBlack prefix collectd.localhost',  # ignore hosts that don't set their hostname properly (implicit substring matrch).
-     'addBlack regex ^foo\..*\.cpu+', # ignore foo.<anything>.cpu.... (regex pattern match)
-     'addAgg sum ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._sum_$1.requests.$2 10 20',
-     'addAgg avg ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._avg_$1.requests.$2 5 10',
-     'addRoute sendAllMatch carbon-default  127.0.0.1:2005 spool=true pickle=false',
-     'addRoute sendAllMatch carbon-tagger sub==  127.0.0.1:2006',  # all metrics with '=' in them are metrics2.0 format for tagger
-     'addRoute sendFirstMatch analytics regex=(Err/s|wait_time|logger)  graphite.prod:2003 prefix=prod. spool=true pickle=true  graphite.staging:2003 prefix=staging. spool=true pickle=true'
+#     'addBlack regex ^foo\..*\.cpu+', # ignore foo.<anything>.cpu.... (regex pattern match)
+#     'addAgg sum ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._sum_$1.requests.$2 10 20',
+#     'addAgg avg ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._avg_$1.requests.$2 5 10',
+#     'addRoute sendAllMatch carbon-default  127.0.0.1:2005 spool=true pickle=false',
+     'addRoute sendAllMatch carbon-default  127.0.0.1:2005 prepend=hello. append=.world spool=true pickle=false',
+#     'addRoute sendAllMatch carbon-tagger sub==  127.0.0.1:2006',  # all metrics with '=' in them are metrics2.0 format for tagger
+#     'addRoute sendFirstMatch analytics regex=(Err/s|wait_time|logger)  graphite.prod:2003 prefix=prod. spool=true pickle=true  graphite.staging:2003 prefix=staging. spool=true pickle=true'
 ]
 
 [instrumentation]
--- a/admin_http.go
+++ b/admin_http.go
@@ -3,13 +3,14 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	assetfs "github.com/elazarl/go-bindata-assetfs"
-	"github.com/gorilla/mux"
-	"github.com/graphite-ng/carbon-relay-ng/aggregator"
 	"net/http"
 	"os"
 	"strconv"
 	"time"
+
+	assetfs "github.com/elazarl/go-bindata-assetfs"
+	"github.com/gorilla/mux"
+	"github.com/graphite-ng/carbon-relay-ng/aggregator"
 )
 
 // error response contains everything we need to use http.Error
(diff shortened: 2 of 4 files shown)


tcolgate:
If you are interested in this, I'll do the work to add ui, tests and benchmarks, and try and quantify the impact of actually using them. There's likely to be more efficient ways of doing both prepend and append too.

tcolgate:
The buffer updates are probably broken and racey right now. And it won't play well with consistent hashing should probably do this on the route rather than destination?

tcolgate:
Fixed prepend/append to be safer and do less allocation

carbon-relay-ng.ini
```diff
@@ -24,12 +24,13 @@ legacy_metric_validation = "strict"
...
-     'addRoute sendAllMatch carbon-tagger sub==  127.0.0.1:2006',  # all metrics with '=' in them are metrics2.0 format for tagger
-     'addRoute sendFirstMatch analytics regex=(Err/s|wait_time|logger)  graphite.prod:2003 prefix=prod. spool=true pickle=true  graphite.staging:2003 prefix=staging. spool=true pickle=true'
+#     'addBlack regex ^foo\..*\.cpu+', # ignore foo.<anything>.cpu.... (regex pattern match)
+#     'addAgg sum ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._sum_$1.requests.$2 10 20',
+#     'addAgg avg ^stats\.timers\.(app|proxy|static)[0-9]+\.requests\.(.*) stats.timers._avg_$1.requests.$2 5 10',
+#     'addRoute sendAllMatch carbon-default  127.0.0.1:2005 spool=true pickle=false',
+     'addRoute sendAllMatch carbon-default  127.0.0.1:2005 prepend=hello. append=.world spool=true pickle=false',
+#     'addRoute sendAllMatch carbon-tagger sub==  127.0.0.1:2006',  # all metrics with '=' in them are metrics2.0 format for tagger
+#     'addRoute sendFirstMatch analytics regex=(Err/s|wait_time|logger)  graphite.prod:2003 prefix=prod. spool=true pickle=true  graphite.staging:2003 prefix=staging. spool=true pickle=true'
```
Dieterbe: the change that you commit for this file should not uncomment all the existing ones. maybe just add your unew entry and put a comment with what it does (protip: `git add -p` and `git rebase -i origin/master` && git push -f)

Dieterbe:
In principle i have nothing against a feature like this, as long as it's not to intrusive, which doesn't seem to be the case. and as long as there's a good use case. can you explain the use case? What problem does it solve?

> And it won't play well with consistent hashing should probably do this on the route rather than destination?

no other relay supports prepend/append so far right? at least not the official python ones?
If so, it's up to us decide whether we want the CH logic to take into account the original metric name (ie. do prepend/append in the destination or in the route right before we send to the destinations), or whether the CH logic should take the modified metric names for its routing logic (in which case it should be in the route before we pick the destinations).
I can see how both can make sense, i guess maybe it depends on how you think of what the append/prepend represents.  So again, use case seems relevant here.

we could also implement this as a processing step, similar to how aggregators and blacklist work.
then it would of course apply to all metrics and not on a per-route basis, which seems less useful. (or prepend/append would only apply if a certain per append/prepend rule matcher matches, but that matcher would probably be the same as the one in the route anyway, and matching twice seems less efficient.)

carbon-relay-ng.ini
tcolgate: Sorry about that, I didn't mean to commit that at all. Will rebase it out.

tcolgate:
> In principle i have nothing against a feature like this, as long as it's not to intrusive, which doesn't seem to be the case. and as long as there's a good use case. can you explain the use case? What problem does it solve?

We need something like this for routing metrics to dataloop.io. They require all graphite metrics to be prefixed with an agent ID. We have some internal code for doing this, but it seemed like a nice to thing to contribute here too. Strictly we only need the prefix support, I'm not sure how the suffix case would be useful, but it seemed like a natural addition.

> If so, it's up to us decide whether we want the CH logic to take into account the original metric name 

I've not used graphite much, so am not 100% sure on these things. My concern is that if the prefix is added after the destination is added, then, potentially, the logic that does the read side of things in a clustered carbon setup wont pick the right node anymore. Now, for our use case, I don't really care about that, since we wont be reading these from graphite, (we'll be reading them from dataloop). But it does feel like it would play nicer with regular graphite if it was done in the routes? I may also be misunderstanding how CH works.

> no other relay supports prepend/append so far right? at least not the official python ones?
No, though I did see a node.js one that seemed to.

Dieterbe:
and dataloop has an api point that takes in the graphite protocol?  e.g. you can just configure their address as a destination? how does that work with authentication?

tcolgate:
> and dataloop has an api point that takes in the graphite protocol? e.g. you can just configure their address as a destination? how does that work with authentication?

I think the client ID is used as an auth token from their side. You run the agent to get an initial client/agent ID. You can then use that as a prefix on the metrics you pass to the graphite endpoint to route the data to your account (they have non-carbon data ingestion options too, but that's the easiest one for us to get started). We're only just testing this stuff at the moment.

tcolgate:
I've read up on graphite clustering a little more. It seems that the web side of things doesn't actually care about consistant hashing. it's only needed for picking the destination to write to (querying queries all remote web servers, it doesn't base it on consistant hashing).
  That makes me feel a little less guilty about doing it on the destination. Consistant hashing will still spread the load. It'll end up on the same host it would have before the name was altered. It just wont be on the one you expected it to be on if you re-hash it yourself.
  I think I'll still look at doing it on the route. That seems less likely to cause suprises.

joshmyers:
Hi, any movement on this PR? we are looking to forward metrics onto Hosted Graphite, where each metric needs to be prefixed with an API key. It would be nice to do this in carbon-relay-ng. Thanks

tcolgate:
I did the original work when we were looking at dataloop, we've abandoned
that for locally hosted prometheus.
The PR does work, so you could just use my branch, but I am unlikely to
have time to refactor this any further.

DanCech:
@joshmyers this can now be achieved via regex rewrite rules, see #139

tcolgate:
Might as well close this PR then. (I think there's a case to be made re:
performance, but realistically I don't have the time to support this code)

Dieterbe:
Rewrite rules are global though. This is on route/destination level. Important difference.

Dieterbe:
This approach seems to be gaining more popularity amongst hosted metrics providers, and at my employer (grafanacloud) we will also support this mechanism.
I've been thinking about it a bit more, and think this should be done on the destination level. since hosted metrics providers typically have 1 endpoint to connect to, so doing it on the route level, which is really meant to manage the distribution of a stream across many different endpoints, doesn't make so much sense anymore, as you would want it on just 1 of the endpoints anyway.

I'll see about cleaning this code up. thanks for your initial version @tcolgate .

tcolgate:
@Dieterbe did this get integrated in some form? Should I close this PR, or perhaps people were happy enough with the changes and a rebase might get merged?

tcolgate:
Closing this PR. I'll leave my repo/branch for reference if people find this and are interested.

PR #153, opened 2017-01-25
Concurrent grafana net

woodsaj:
allow grafanNet route to make concurrent connections to tsdb-gw
    
 - adds a "concurrency" config setting for defining the maximum number
      of concurrent http requests that can be made
 - split making http requests into separate goroutines so they dont
      block the main loop.  This allows new metrics to be parsed and
      added to the buffer while the "flush" goroutines are IO blocked.
 - shard metrics into multiple buffers based on concurrency. This ensures
      any delays processing a buffer dont lead to metrics being sent out of
       order.

### DIFF
--- a/route/grafananet.go
+++ b/route/grafananet.go
@@ -39,6 +39,7 @@ type GrafanaNet struct {
 	timeout      time.Duration
 	sslVerify    bool
 	concurrency  int
+	orgId        int
 	writeQueues  []chan []byte
 	shutdown     chan struct{}
 	wg           *sync.WaitGroup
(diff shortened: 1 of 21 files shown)


woodsaj:
@Dieterbe this is ready for review

Dieterbe:
see #114

route/grafananet.go
```diff
@@ -125,41 +132,99 @@ func (route *GrafanaNet) run() {
...
+	for {
+		select {
+		case buf := <-route.buf:
+			route.numBuffered.Dec(1)
+			md := parseMetric(buf, route.schemas)
+			if md == nil {
+				continue
+			}
+			md.SetId()
+			buf = md.KeyBySeries(buf[:0])
```
Dieterbe: interesting to reuse `buf` here for the key. this is a bit confusing since buf was our metric message, but seems safe since parseMetric returned `md` which is decoupled from any of buf's memory.

i think this warrants a comment though.

woodsaj:
@Dieterbe can we get this merged?

Dieterbe:
thank you come again

PR #173, opened 2017-03-30
updated pickle.go to use ogorek.Tuple instead []interface{} type for item and data

isavcic:
It seems that upstream module, kisielk/ogorek, created a custom type `Tuple` to reflect Python's tuple. In order for carbon-relay-ng to use the new ogorek version, I had to switch from `[]interface{}` to `ogorek.Tuple` for `item` and `data`. My apologies for I didn't see this all the way through when I PR'd the new ogorek version.

### DIFF
--- a/input/pickle.go
+++ b/input/pickle.go
@@ -35,6 +35,7 @@ func (p *Pickle) Handle(c net.Conn) {
 	defer c.Close()
 	// TODO c.SetTimeout(60e9)
 	r := bufio.NewReaderSize(c, 4096)
+	log.Debug("pickle.go: entering ReadLoop...")
 ReadLoop:
 	for {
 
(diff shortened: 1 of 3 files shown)


isavcic:
I've tested this manually with some pickles and the Whisper files showed up. Waiting for more prod data to arrive.

isavcic:
But on the other hand, some of the non-manually generated metrics are not being updated, I'm going to perform a tcpdump and enable the Python carbon-relay for pickled metrics again, to verify they're getting updated by it. More testing needed.

isavcic:
To state again: with these changes, I see no errors whatsoever, the metrics I'm sending manually via a simple Python code are received properly, yet some prod data is missing, while it's updated properly via Python carbon-relay.

Example of the test which goes through:

```python
#!/usr/bin/python
import pickle
import time
import struct
import socket

list = [('pickletest.foobar1', (time.time(), 42)), ('pickletest.foobar2', (time.time(), 42))]

print list
payload = pickle.dumps(list)
header = struct.pack("!L", len(payload))
message = header + payload

def main():
	s = socket.socket()
        s.connect(('graphite', 2004))
        s.sendall(message)
        s.close()
main()
```

isavcic:
I see the manually sent metrics in the debug log, loud and clear. The prod senders connect to the TCP port, but it's like they don't want to send to carbon-relay-ng, they just remain connected, sending no data, thus nothing appears in debug log as well. Weird...

isavcic:
I will expand upon this as time permits and create another pull request. Disregard this.

PR #337, opened 2018-12-27
Fix init script and RPM build for CentOS 6

mbethke:
The RPM for CentOS 6 isn't very useful as is because it installs and init script that looks for the daemon in the wrong place (and thus exists silently without doing anything) and tries to source files that don't exist on CentOS 6, and it doesn't create an account for privilege separation nor the default directories for logs etc.

This fixes a lot of these issues using a few pre- and post-install RPM scripts and a basically rewritten init script.

I'm not completely happy with slapping everything in `examples/` and the way this isn't shared with the other RPM-based distributions so while I hope this will make it upstream I'd be happy to improve on it if anyone has suggestions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,14 +103,17 @@ rpm: build-linux
 	rm -rf redhat
 
 rpm-centos6: build-linux
-	mkdir build/centos-6
-	install -d redhat/usr/bin redhat/usr/share/man/man1 redhat/etc/carbon-relay-ng redhat/etc/init redhat/etc/init.d
+	mkdir -p build/centos-6
+	install -d redhat/usr/bin redhat/usr/share/man/man1 redhat/etc/carbon-relay-ng \
+		redhat/etc/init redhat/etc/init.d \
+		redhat/var/lib/carbon-relay-ng redhat/var/log/carbon-relay-ng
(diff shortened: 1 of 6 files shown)


npazosmendez:
Hello! I'm doing some housekeeping to this repo and I notice this old PR, sorry we didn't get to this earlier. If the CentOS 6 build is indeed broken, then I think it makes sense to include this fix, though I don't have a way to test it. If you still wish to work on this, could you please:

- Rebase this PR
- Sign the CLA
- Confirm that this is working :slightly_smiling_face: 


Thank you for your contribution

mbethke:
Hi Nico! Thanks for the effort to understand my garbled description (I blame autocorrect ^^) but I don't even have a CentOS-6 machine to test this on any more either. With C6 long EOL, feel free to reuse any bits that might be useful for other RPM distros but otherwise I guess we can just close this.

npazosmendez:
Thanks for the quick response, I was not expecting it given the age of this PR :slightly_smiling_face: . 

>  With C6 long EOL, feel free to reuse any bits that might be useful for other RPM distros but otherwise I guess we can just close this.

Sounds good. I will close this PR for now, we can come back to it in the future as necessary. Please feel free to reopen at any time. Cheers!

PR #393, opened 2020-02-03
Add support for passing grafana.net credentials via environment variables

lukaszkorecki:
It doesn't expose all config options to be set via env vars, like here: https://github.com/grafana/carbon-relay-ng/issues/192 but helps with configuring grafana.net route at least.

I've extracted how the config file is read, so there is a centralized place for adding support for more environment variables. Technically, I could have used [`os.ExpandEnv`](https://golang.org/pkg/os/#ExpandEnv) but it didn't work in tests and might be a bit too flexible. On the other hand it would allow specifying multiple credentials for the same type of a route, e.g. using variables like `GRAFANA_1_API_KEY` etc.

It's been a while since I wrote any Go code - please let me know if I need to fix anything 👍

### DIFF
--- a/cmd/carbon-relay-ng/carbon-relay-ng_test.go
+++ b/cmd/carbon-relay-ng/carbon-relay-ng_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -341,6 +342,23 @@ func BenchmarkTableDisPatchAndEndpointReceive(b *testing.B) {
 	tE.Close()
 }
 
+func TestConfigHostVarInterpolation(t *testing.T) {
+	template := []byte(`host = "${HOST}"`)
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		t.Fatal("failed to parse hostname", err)
+	}
+	expected := fmt.Sprintf(`host = "%s"`, strings.Split(hostname, ".")[0])
+
+	ioutil.WriteFile("/tmp/config.example.toml", template, 0644)
+	config := readConfigFile("/tmp/config.example.toml")
+
+	if config != expected {
+		t.Errorf("Expected interpolated config %s but got %s", expected, config)
+	}
+}
+
 func TestConfigEnvVarInterpolation(t *testing.T) {
 	os.Setenv("GRAFANA_NET_ADDR", "foo.com")
 	os.Setenv("GRAFANA_NET_API_KEY", "wow")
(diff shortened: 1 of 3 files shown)


lukaszkorecki:
Bumping this to mention that we had a version of carbon-relay-ng with this patch, in production,  sending data to Grafana Cloud with no issues whatsoever 🙂

cmd/carbon-relay-ng/carbon-relay-ng_test.go
```diff
@@ -339,3 +340,41 @@ func BenchmarkTableDisPatchAndEndpointReceive(b *testing.B) {
...
+host = "${HOST}"
+# [[route]]
+key = 'grafanaNet'
+type = 'grafanaNet'
+addr = "${GRAFANA_NET_ADDR}"
+apikey = "${GRAFANA_NET_USER_ID}:${GRAFANA_NET_API_KEY}"
+`)
+
+	expected_template := `
+host = "work"
```
fitzoh: This test fails for me locally, I'm guessing `work` is your hostname :)

We should probably drop the `HOST` env var since it's not actually used.

I'm a little torn on if we should remove `host` from this test altogether or substitute in the host name dynamically.

lukaszkorecki: Ha, good catch - I left the support for the HOST variable as it was there originally - I can of course update the test to get the value from the current environment to make it reproducible.

lukaszkorecki: @fitzoh Updated!

fitzoh: Awesome!
I pushed a small tweak in 1070d4389adb4b1b8b63a47acfe0d73ee2bd5071 to split the hostname + env var interpolation into separate tests and accounted for the way it splits on `.`

LGTM, I'll poke the team and see if I can get another quick +1

fitzoh:
Thanks for the contribution! (and sorry for the delay)

lukaszkorecki:
Awesome! Do you know, roughly when this version will be available in the public docker hub?

fitzoh:
https://hub.docker.com/r/raintank/carbon-relay-ng/tags?page=1&ordering=last_updated

It's available now @ `raintank/carbon-relay-ng:0.12.0-27-gd26021e` or `raintank/carbon-relay-ng:master`... not sure on the timeline of an actual release

oscherler:
This feature is pretty confusing. It looks like you can use environment variables in the config file, and there’s an *example* on how to do so.

But in reality, there are exactly four environment variables you can use, and they must be named exactly like they are named in the *example*, and so the *example* is not an example but the exact only way you can use environment variables.

I thing a warning should be added to the documentation, because I just used environment variable expansion in my Telegraf config, and it worked as expected with whatever variable names I chose to use, and then when I wanted to do the same for carbon-relay-ng, I had to read the code to understand why it was not working.

PR #411, opened 2020-03-31
Testing to see if this will launch acceptance. 3

hdost:


woodsaj:
@hdost I think the problem is with the PR you are creating. I notice that the branch your PR is targeting is "grafana:master", when it should just be "master".  Are you creating your PR through Github, or through some other tool?

hdost:
I am creating it through github. The reason it displays this way is because it's a Cross Repo PR

PR #445, opened 2020-10-30
upgrade sarama to v1.23.0

### LINKED ISSUE #443: Update sarama to v1.23.0
The sarama dependency should be updated to v1.23.0 (same version as metrictank 1.0).

This will allow users of carbon-relay-ng to use the same versions of kafka as in metrictank which reduces the conversions made by kafka.

carbon-relay-ng currently uses sarama 1.19.0 which supports kafka up to version 2.0.0, metrictank with sarama 1.23.0 supports kafka up to version 2.3.0.

If the producers and consumers work on different versions of sarama / the used kafka version is not supported by one them kafka has to do conversions which reduces the performance of kafka and increases the load.
https://medium.com/@jiangtaoliu/a-lesson-learning-of-kafka-message-conversions-4a9a091759a5

D3luxee:
This will fix #443 

`go get github.com/Shopify/sarama@v1.23.0`


Dieterbe (approved):
lgtm

PR #479, opened 2021-11-12
Build Linux packages for multiple architectures

legoscia:
Currently, when building a Linux package on e.g. an ARM64 machine, the
package produced will be marked as an ARM64 package (because of the
"-a native" flag passed to fpm), but will contain an AMD64
executable (because the build-linux target sets GOARCH
unconditionally).

With this change, the desired architecture can be changed by setting
the LINUX_PACKAGE_GOARCH variable, so that the package will always
match the binary it contains. Also, the "packages" target builds
packages for both AMD64 and ARM64.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine
 RUN apk --update add --no-cache ca-certificates
-ADD carbon-relay-ng /bin/
+ADD carbon-relay-ng-linux-amd64 /bin/carbon-relay-ng
 VOLUME /conf
 ADD examples/carbon-relay-ng.ini /conf/carbon-relay-ng.ini
 RUN mkdir /var/spool/carbon-relay-ng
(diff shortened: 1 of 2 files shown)


legoscia:
I added some commits to make the CircleCI package build pass - not sure if they are useful, or perhaps more appropriate for a separate pull request.

aboroska:
Heads up @fionaliao that this PR enables building for Arm architecture. Are there any plans to review it?

fionaliao:
@aboroska I'll review this, sorry for the long delay.

fionaliao:
> I added some commits to make the CircleCI package build pass - not sure if they are useful, or perhaps more appropriate for a separate pull request.

I think this has been fixed by https://github.com/grafana/carbon-relay-ng/pull/497. 

Also could you rebase these changes?

Other than that, the PR looks good :+1:

fionaliao (approved):

PR #485, opened 2022-02-16
Make aggregationFile an optional setting

fionaliao:
aggregationFile is currently a mandatory setting for configuring a GrafanaNet route. This PR updates it to be optional. Making this optional simplifies the process of migrating to Grafana Cloud Graphite V5.

If aggregationFile is not set, the most recent aggregation file uploaded to Grafana Cloud will be used instead. If no aggregation files have been uploaded, the default Grafana Cloud aggregations will be applied.

This is a **breaking change** for the `addRoute grafanaNet ...` command, as `aggregationFile=` needs to be prepended to the aggregation file value since it's now an optional parameter. 

# Testing
Built a local Docker image for my branch and configured it to push to Grafana Cloud. Checked that ignoring `aggregationFile` doesn't error, and also checked that setting `aggregationFile` causes the config to be pushed to Grafana Cloud.


imperatives/imperatives.go
```diff
@@ -507,20 +510,27 @@ func readAddRouteGrafanaNet(s *toki.Scanner, table table.Interface) error {
...
 	if err != nil {
 		return errFmtAddRouteGrafanaNet
 	}
 
+	t = s.Next()
+
 	for ; t.Token != toki.EOF; t = s.Next() {
 		switch t.Token {
+		case optAggregationFile:
+			t = s.Next()
```
jesusvazquez, line 523: Why do you need to do this here?

fionaliao: The value of `t.Token` when we enter the for loop will be the name of the optional setting (in this case `aggregationFile=`). We need to move to the next token to get the value of the setting.

jesusvazquez (approved):
Overall looks good to me :muscle:

route/grafananet_test.go
```diff
@@ -63,33 +63,25 @@ func TestNewGrafanaNetConfig(t *testing.T) {
 		{otherFile.Name(), true},
 		{schemasFile.Name(), false},
 	}
-	aggregationFileOptions := []option{
-		{"", true},
-		{"some-path-that-definitely-will-not-exist-for-carbon-relay-ng", true},
-		{otherFile.Name(), true},
-		{aggregationFile.Name(), false},
-	}
```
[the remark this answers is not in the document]
fionaliao: Re-added the test cases in https://github.com/grafana/carbon-relay-ng/pull/485/commits/4cad65f3405056163a00399b785ba5c028a64c0c

(The original PR removed all references to the aggregation file in the `NewGrafanaNetConfig` function, so these tests weren't useful anymore. Since the aggregation file argument has been re-added to this function, these test cases were re-added too)

cfg/table.go
```diff
@@ -210,12 +210,13 @@ func InitRoutes(table table.Interface, config Config, meta toml.MetaData) error
 			table.AddRoute(route)
 		case "grafanaNet":
 
-			cfg, err := route.NewGrafanaNetConfig(routeConfig.Addr, routeConfig.ApiKey, routeConfig.SchemasFile, routeConfig.AggregationFile)
+			cfg, err := route.NewGrafanaNetConfig(routeConfig.Addr, routeConfig.ApiKey, routeConfig.SchemasFile)
```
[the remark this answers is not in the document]
fionaliao: I reverted this change in https://github.com/grafana/carbon-relay-ng/commit/4cad65f3405056163a00399b785ba5c028a64c0c - now `NewGrafanaNetConfig` takes aggFile as a parameter, and if it's not the empty string, tries to validate it. Like you mention, in `imperatives.go` we still need to manually validate and set aggregationFile. 

My initial thought was to directly set the aggregationFile field after the config had been created, since it's an optional field, and that's how all the other optional fields are set. However, that could lead to accidentally forgetting to validate the field before we try and create a new route with the config like I originally did here, so I applied your suggestion. 

I think it's still not ideal since there are two paths for setting the aggregationFile and both need to do validation. A future improvement could be to take in option functions in the config constructor. That way setting the optional fields won't require the config instance to be created before they can be set.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * adopt the term "blocklist" for blocking metrics.  The existing config options "blacklist" and commands "addBlack" are replaced
   with "blocklist" and "addBlock" respectively.  The old option and command will keep working for the time being, but users are recommended
   to update their configuration because the old options will be removed at the next major release. #473
+* grafanaNet route: make publishing storage-aggregation.conf optional. This is a breaking change for the `addRoute grafanaNet ...` command, as `aggregationFile=` now needs to be prepended to the aggregation file value. #485
```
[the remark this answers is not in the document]
fionaliao: Makes sense, clarified in https://github.com/grafana/carbon-relay-ng/pull/485/commits/75b30835e304f29e42936d1f357ce7215340e6fb

fionaliao:
>  i recommend you grep for aggregations throughout the repo to find what needs updating.

This has been done, mainly in these two commits: https://github.com/grafana/carbon-relay-ng/pull/485/files/0379100cd31a361bbf0fcb3b7da04e9f7219d88a..414da429ba9bd73f696938cf7c843d8b0faa0db3

> jsonnet (remove the default)

I decided to keep the aggregation file config in the jsonnet, but only add the file to the config map if `$._config.storage_aggregation` was not null:

https://github.com/grafana/carbon-relay-ng/blob/414da429ba9bd73f696938cf7c843d8b0faa0db3/jsonnet/lib/carbon-relay-ng/crng.libsonnet#L14-L32

The reasoning was that having the aggregation file and configmap set up is a useful reference if you want to configure your aggregations, but there are also simple and clear steps you can take if you don't want to send an aggregation file.

route/grafananet.go
```diff
@@ -163,7 +171,7 @@ func NewGrafanaNet(key string, matcher matcher.Matcher, cfg GrafanaNetConfig) (R
 	if cfg.AggregationFile != "" {
 		aggregation, err = conf.ReadAggregations(cfg.AggregationFile)
 		if err != nil {
-			return nil, fmt.Errorf("NewGrafanaNet: could not read aggregationFile %q: %s", cfg.AggregationFile, err.Error())
+			return nil, err
```
Dieterbe, line 174: why? and why do we do it differently here than in NewGrafanaNetConfig() ?

Dieterbe:
> I decided to keep the aggregation file config in the jsonnet

ok. does the default match what we have in cloud v5?

Dieterbe (changes requested):

route/grafananet.go
fionaliao: This just wraps the original  (pre-PR) `ReadAggregations(...)` call in an if-else because it's possible that no file will be supplied. 

I see you're commenting against a diff within two commits of the PR though:
image
In an earlier commit, I added the `"NewGrafanaNet:..."` error since I'd removed the aggregation file parameter from `NewGrafanaNetConfig()`. After the suggestion to keep the aggregation file parameter, I reverted that change and then reverted the error in this method back to what it was originally - just returning the error from reading the aggregations.

fionaliao:
> ok. does the default match what we have in cloud v5?

It didn't, but I updated the examples to match in https://github.com/grafana/carbon-relay-ng/pull/485/commits/e4294f0ec7a6fd9fc09b51a6afefe38c4910510b

Dieterbe (approved):
nice work, fiona. thanks!

PR #487, opened 2022-02-28
Update Jsonnet lib to work with latest Tanka version

fionaliao:
The default k8s library used by Tanka was changed in https://github.com/grafana/tanka/pull/563. This meant the [k8s instructions](https://github.com/grafana/carbon-relay-ng/blob/master/docs/deploying-on-k8s.md) didn't work with the latest version of Tanka (v0.20); I was seeing an error like:
```
$ tk show environments/default/
Error: evaluating jsonnet: RUNTIME ERROR: Attempt to use super when there is no super class.
        /.../crng-tanka-check/vendor/github.com/jsonnet-libs/k8s-libsonnet/1.20/_custom/apps.libsonnet:35:7-12 function <anonymous>
        /.../crng-tanka-check/vendor/ksonnet-util/grafana.libsonnet:45:9-57    function <anonymous>
        /.../crng-tanka-check/vendor/carbon-relay-ng/crng.libsonnet:62:5-94    object <anonymous>
        Field "carbon_relay_ng_deployment"
        Field "crng"
        During manifestation
```

This PR updates the carbon-relay-ng jsonnet lib so it'll work with the updated k8s library, and also makes a small update to the k8s doc. 

# Testing
Followed the k8s instructions to set up the Tanka environment, ran `tk show environments/default`, which worked after I'd made changes to the jsonnet library.

### DIFF
--- a/docs/deploying-on-k8s.md
+++ b/docs/deploying-on-k8s.md
@@ -36,8 +36,8 @@ local crng = import 'carbon-relay-ng/crng.libsonnet';
   crng: crng {
     _config+:: {
       namespace: 'mynamespace',
-      crng_route_host: 'https://tsdb-1-<instance name>.hosted-metrics.grafana.net/metrics',
-      crng_user_id: 'api_key',
+      crng_route_host: '<graphite endpoint>/metrics',
+      crng_user_id: '<user id>',
       crng_api_key: '<base64 api key>',
     },
     _images+:: {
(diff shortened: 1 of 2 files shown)


docs/deploying-on-k8s.md
```diff
@@ -36,8 +36,8 @@ local crng = import 'carbon-relay-ng/crng.libsonnet';
   crng: crng {
     _config+:: {
       namespace: 'mynamespace',
-      crng_route_host: 'https://tsdb-1-<instance name>.hosted-metrics.grafana.net/metrics',
```
fionaliao, line 39: I updated this value as the cloud graphite v5 endpoints don't follow the `tsdb-1-<instance name>.hosted-metrics.grafana.net` pattern.

Dieterbe (approved):

PR #492, opened 2022-08-22
Pp 9940 no op change

jfharden:
The post-merge workflow didn't have the checkout action approved and you can't re-run workflows that fail so early in the run.

This no-op change will force the workflow to run again.


jfharden:
I'm so sorry, this was against the wrong repo.

I know this is the second time I've opened a PR against the wrong repository today. I'll make sure it doesn't happen again

PR #498, opened 2022-12-22
Use go releaser to publish binaries on new tags

jesusvazquez:
This PR updates circle ci config to use goreleaser to publish binaries on github releases whenever a new tag is created.


.circleci/config.yml
```diff
@@ -1,4 +1,38 @@
...
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*(-.*)*/
+      - github_binaries:
+          requires:
+            - test
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*(-.*)*/
```
jesusvazquez, line 34: Had to repeat the filtering in both jobs because there is no way to filter an entire workflow

.circleci/config.yml
```diff
@@ -96,21 +130,12 @@ jobs:
...
-            - package
+
+  github_binaries:
+      docker:
+        - image: cimg/go:1.18
+      steps:
+        - checkout
+        - run: curl -sfL https://goreleaser.com/static/run | bash
+        # For dry runs use the following and comment the above
+        # - run: curl -sfL https://goreleaser.com/static/run | bash -s -- --rm-dist --skip-publish --snapshot
```
jesusvazquez, line 141: Took me a few minutes to find out how to do a dry run so I think this comment adds value for future us if we run into any issue

.goreleaser.yml
```diff
@@ -0,0 +1,25 @@
...
+    main: ./cmd/carbon-relay-ng
+    binary: carbon-relay-ng
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - amd64
```
jesusvazquez, line 12: amd64 only arch supported today because we ran into issues when compiling to arm due to our metrictank dependency

npazosmendez (approved):

PR #506, opened 2023-02-21
Deploy new packages when tag is added

fionaliao:
If you merge a new release into master and then add a new tag, the publish_release workflow fails ([example](https://app.circleci.com/pipelines/github/grafana/carbon-relay-ng/300/workflows/97c46183-8ee6-42a8-bcdb-2fc0201dc304/jobs/2779)). This is because packages aren't being deployed when a new tag is added. This PR adds the build/package/deploy steps to the publish_release workflow so correct packages are being deployed before the publish step.

Also fixed an [issue](https://app.circleci.com/pipelines/github/grafana/carbon-relay-ng/310/workflows/4b743e84-ea34-424c-9583-effcff1aba5a/jobs/2835) with packaging - after https://github.com/grafana/carbon-relay-ng/pull/479 the package step started failing because it was trying to build a binary before packaging it. The fix here was to make sure both the arm64/amd64 binaries are being created in the build step so the package step doesn't try to build a new binary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,9 +24,39 @@ workflows:
               ignore: /.*/
             tags:
               only: /v[0-9]+(\.[0-9]+)*(-.*)*/
+      - build:
+          requires:
+            - test
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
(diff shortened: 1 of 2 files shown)


jesusvazquez (approved):
LGTM, try adding arm arch to goreleaser as well

PR #544, opened 2024-03-12
Bump go version to 1.21

jesusvazquez:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grafana/carbon-relay-ng
 
-go 1.20
+go 1.21.8
```
ywwg, line 3: should we just say go 1.21 and that will pick up whatever the latest go 1.21 is?

ywwg (approved):

PR #547, opened 2024-03-26
fix(pickle): allow big.Int as value

### LINKED ISSUE #546: pickle.go: Unrecognized type *big.Int for value
When sending big numbers to carbon-relay-ng it fails with the error message: 
```
[ERROR] pickle.go: Unrecognized type *big.Int for value
```

You can reproduces the error with this script:
```python
import socket
import pickle
import time
import struct
import math


def send_to_graphite(metric, value, timestamp=None, host="localhost", port=2004):
    if timestamp is None:
        timestamp = int(time.time())

    # Construct the data as a list of tuples
    data = [(metric, (timestamp, value))]

    # Pickle the data
    payload = pickle.dumps(data)

    try:
        # Create a socket connection to Graphite server
        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        sock.connect((host, port))
        # Send the pickled data
        size = struct.pack("!L", len(payload))
        sock.sendall(size)
        sock.sendall(payload)
        sock.close()
        print(f"Sent {metric} = {value} to Graphite at {host}:{port}")
    except Exception as e:
        print(f"Error sending data to Graphite: {e}")


if __name__ == "__main__":
    large_int = int(math.pow(10, 20))  # A large integer exceeding regular int range

    send_to_graphite("my.test.metric", large_int)
```

I've already fixed it by adding `(*big.Int)` to the allowed datatypes. `(*big.Int)` was already allowed for the timestamp.
I'll create a PR.

ljurk:
Closes #546

### DIFF
--- a/input/pickle.go
+++ b/input/pickle.go
@@ -157,7 +157,7 @@ func (p *Pickle) Handle(c io.Reader) error {
 			switch data[1].(type) {
 			case string:
 				value = data[1].(string)
-			case uint8, uint16, uint32, uint64, int8, int16, int32, int64:
+			case uint8, uint16, uint32, uint64, int8, int16, int32, int64, (*big.Int):
 				value = fmt.Sprintf("%d", data[1])
 			case float32, float64:
 				value = fmt.Sprintf("%f", data[1])


jesusvazquez (approved):
LGTM thanks for your contribution

Issue #551, opened 2024-08-07
The dependency toki isn't licensed


asponzan:
In reviewing the dependencies for carbon-relay-ng we discovered that the upstream of toki, https://github.com/taylorchu/toki, doesn't have a declared license. Does the project have explicit permission to use this code?

asponzan:
Understanding that there hasn't been movement on https://github.com/taylorchu/toki/issues/2 are there plans to mitigate this copyright issue?

npazosmendez:
This should be resolved now, thank you for your patience!  🙂 See https://github.com/taylorchu/toki/commit/36c157adcc65b5cfec7781167acbfb4ea240d469

PR #553, opened 2024-09-24
Remove toki

mbells:
Remove the toki module, replaced with a rewritten tokre. This is a tokenizer based on regular expressions.
The API is similar, but not the same.
The tokre module is pulled into the same repo, as there are not many users of the original code.
Fixes issue https://github.com/grafana/carbon-relay-ng/issues/551

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,6 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/streadway/amqp v0.0.0-20170521212453-dfe15e360485
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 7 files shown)


npazosmendez:
Thank you for working on this and sorry for the delay on our end. We had an internal discussion about how to proceed about this. Fortunately, toki's maintainer just added an MIT license to the project, which I believe resolves your concern: https://github.com/taylorchu/toki/commit/36c157adcc65b5cfec7781167acbfb4ea240d469

Having that said, I will close this PR.

mbells:
Thanks. We're more than happy that it was resolved by updating the license of the original package.

PR #571, opened 2024-12-04
docker: build using distroless image

jesusvazquez:
This PR updates the dockerfile to use alpine as the builder image but base the final carbon-relay-ng image on a distroless image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,15 @@
-FROM alpine
+FROM alpine as builder
 RUN apk --update add --no-cache ca-certificates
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,15 @@
-FROM alpine
+FROM alpine as groupbuilder
```
npazosmendez, line 1: I'd name this `builder` maybe

npazosmendez (approved):
LGTM, just a minor comment

Dockerfile
jesusvazquez: renamed and rebased, good call

PR #582, opened 2025-01-22
grafananet: fix FlushSize metric

narqo:
In the `grafananet`router, the definition of the `FlushSize` metrics says its unit is Bytes. But the router's code observes the actual number of metrics sent out, which is also tracked by the `unit=Metric.direction=out` counter.

This PR fixes the router to track the number of bytes sent out to the destination with the request. This behaviour looks aligned with other routers, e.g. see `KafkaMdm`.

### DIFF
--- a/route/grafananet.go
+++ b/route/grafananet.go
@@ -340,7 +340,7 @@ func (route *GrafanaNet) retryFlush(metrics []*schema.MetricData, buffer *bytes.
 		// re-instantiate body, since the previous .Do() attempt would have Read it all the way
 		req.Body = ioutil.NopCloser(bytes.NewReader(body))
 	}
-	log.Debugf("GrafanaNet sent %d metrics in %s -msg size %d", len(metrics), dur, len(metrics))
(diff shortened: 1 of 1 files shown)


route/grafananet.go
```diff
@@ -338,9 +340,9 @@ func (route *GrafanaNet) retryFlush(metrics []*schema.MetricData, buffer *bytes.
 		// re-instantiate body, since the previous .Do() attempt would have Read it all the way
 		req.Body = ioutil.NopCloser(bytes.NewReader(body))
 	}
-	log.Debugf("GrafanaNet sent metrics in %s -msg size %d", dur, len(metrics))
+	log.Debugf("GrafanaNet sent %d metrics in %s -msg size %d", len(metrics), dur, len(metrics))
```
jesusvazquez, line 343: Aren't we printing the number of metrics twice in this message now?

narqo: Good call — fixed

jesusvazquez (approved):
LGTM!

PR #590, opened 2025-03-31
ci: fix packagecloud and docker non-testing builds

npazosmendez:
The version in `build/version.txt` does not contain the leading `v`, because of the `sed 's/^v//' ` here:

https://github.com/grafana/carbon-relay-ng/blob/7d146a931743a9f5f72d3d0b9eebcab7136e1cfc/.github/workflows/ci.yaml#L55

So this regex matching was never succeeding, and we were always pushing to `testing` for packagecloud, and also never `latest` for docker.


krajorama (approved):
LGTM

PR #604, opened 2025-04-30
fix(actions): do not pass secrets undefined on receiver side

krajorama:
Related to https://github.com/grafana/db-upstream-squad/issues/717

We got
```
Invalid workflow file: .github/workflows/dependabot-reviewer.yml#L12
The workflow is not valid. .github/workflows/dependabot-reviewer.yml (Line: 12, Col: 21): Invalid secret, GITHUB_TOKEN is not defined in the referenced workflow. .github/workflows/dependabot-reviewer.yml (Line: 13, Col: 30): Invalid secret, DEPENDABOTREVIEWER_ID is not defined in the referenced workflow.
```


alexgreenbank (approved):
stamp

PR #746, opened 2026-05-29
ci: smoke test built docker images

npazosmendez:
Adds a smoke test that runs each built image (main and k8s) in CI. As the script says, it confirms every listener logs that it is up, then sends a metric and verifies the routing table counter is incremented.

Piggybacking cleanup: delete `build_docker.sh`. Centralize version computation behind `make print-version`, specific tagging is handled directly in the CI definition.


smoke_test.sh
```diff
@@ -0,0 +1,55 @@
...
+# Retry a command quietly until it succeeds, giving up after ~30s.
+retry() {
+    for _ in $(seq 30); do
+        "$@" >/dev/null 2>&1 && return 0
+        sleep 1
+    done
+    return 1
+}
+
+docker run -d --name "$container" -e CRNG_LOG_LEVEL=info \
```
krajorama, line 31: What's stopping the container?

smoke_test.sh
```diff
@@ -0,0 +1,55 @@
...
+# 2. A metric sent over the line protocol increments the "direction=in" counter in the routing table
+in_count() {
+    curl -fsS "http://127.0.0.1:${http_port}/debug/vars2" \
+        | grep -oE '"[^"]*\.direction_is_in"[: ]+[0-9]+' | grep -oE '[0-9]+$'
+}
+printf 'carbon.relay.ng.smoke.test 1 %s\n' "$(date +%s)" \
+    > "/dev/tcp/127.0.0.1/${line_port}" || die "could not reach the line listener"
+ingested() { [ "$(in_count)" -gt 0 ]; }
+retry ingested || die "metric was sent but never counted"
+
```
krajorama, line 54: nit: let's dump the docker logs on success as well , would be cool to see 😊

krajorama (commented):
Looks awesome, way better than I imagined.
I think it would make sense to also dump the logs on success, just

smoke_test.sh
npazosmendez: The `cleanup` trap on `EXIT` runs `docker rm -f` (line 13), which kills and removes the container when the script exits, on both success and failure.

krajorama (approved):
LGTM

PR #747, opened 2026-05-29
docs: update for frozen Docker image tags

npazosmendez:
Updates docs and examples to reflect the Docker image publishing changes from PRs #742-746. The mutable `latest`, `main`, and `master` tags are now frozen. Users should pin to release versions or use immutable `main-<short-sha>` tags instead.


krajorama (approved):
lgtm

PR #775, opened 2026-06-09
build: deprecate packagecloud packages and old distros builds

npazosmendez:
Now that we've verified that our apt/rpm.grafana.com packaging is working as intended, announce deprecation of Packagecloud, in various places:

- Warning when installing from it, for the next release to be done immediately.
- The changelog
- The installation docs, which are also updated to reference the new repository. Instructions match what we use for other packages we publish there, like e.g. [Grafana](https://grafana.com/docs/grafana/latest/setup-grafana/installation/debian/?pg=moving-to-packages-grafana-com&plcmt=in-text).

Additionally, deprecate packages for the very old distros we were still "supporting". We were already deprecating this by not building it for apt/rpm.grafana.com, but I am additionally deprecating them for this last Packagecloud release, because I cannot test them. It is likely that they were not working anyway, see https://github.com/grafana/carbon-relay-ng/pull/337 .

PR #776, opened 2026-06-09
build: remove Packagecloud publishing

npazosmendez:
Packagecloud was deprecated in #775.DEB/RPM packages are now published to apt.grafana.com / rpm.grafana.com via goreleaser nfpm. 

We just did one last Packagecloud release, and now we don't need these anymore.


krajorama (approved):
LGTM

